PR #4515, opened 2024-12-16
Revert "perf: Platform global"

feng626:
Reverts jumpserver/lina#4514

### DIFF
--- a/src/router/console/assets.js
+++ b/src/router/console/assets.js
@@ -2,8 +2,6 @@ import i18n from '@/i18n/i18n'
 import empty from '@/layout/empty'
 import XPackRoutes from './xpack'
 
-const globalSubmenu = () => import('@/layout/globalOrg.vue')
-
 export default [
   {
     path: 'assets',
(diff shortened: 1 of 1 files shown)


src/router/console/assets.js
```diff
@@ -295,7 +293,7 @@ export default [
   },
   {
     path: 'platforms',
-    component: globalSubmenu,
+    component: empty,
     meta: {
       permissions: ['assets.view_platform'],
       resource: 'platform',
```
BaiJiangJie, line 299: There are no immediate observations of irregularities or issues with the provided code. The proposed changes involve:

- Refactoring the `globalSubmenu` into an inline import statement instead of creating a separate namespace entry.

No significant optimizations/suggestions necessary for this change. It aims to improve readability and efficiency slightly but without affecting functionality significantly.

The current difference doesn't seem pertinent enough to warrant further investigation; it's merely a matter of style preference for organizing imports.

Therefore, any additional observations or improvements that might be suggested could focus on these small refinements rather than major structural differences between versions 2 and 3 (as requested by setting the knowledge cutoff point to 2021-09-01).

w940853815 (approved):

PR #4618, opened 2025-02-07
Fixed: Export Button Logic Differentiation

ZhaoJiSen:


src/components/Table/ListTable/TableAction/RightSide.vue
```diff
@@ -155,13 +155,20 @@ export default {
...
        *  那如果首次点击的树节点，那么此时 url 就会被确定，后续点击的树节点，那么 url 就不会
        *  改变了
+       *
+       *  主要是在资产列表中，切换节点的时候
        */
-      return Object.assign({}, this.exportOptions, { url: this.tableUrl })
+      return Object.assign(this.exportOptions, { url: this.tableUrl })
     }
   },
   methods: {
```
BaiJiangJie, line 174: The provided code seems to define a Vue component that exports an asset account list page. The method `exportOptions` has been updated from using the `partialRight` function incorrectly with `assignIfNot` and reusing it with `Object.assign` instead.

The main issue is related to URL assignment when navigating assets lists between nodes or after exporting. Currently, if there's no explicit change on navigation or export options, subsequent calls to `/accounts/account-secrets/` will not lead to different URLs. This needs modification according to the current scenario:

- After successfully switching back to the tree view mode (account details), you may want to adjust how exported URLs are determined based on the user interaction. For instance:
  ```javascript
  exportDefaultMethods() {
    this.importTreeNode(false);
```

In general, these issues should be addressed to ensure consistent behavior across both views while also improving error handling for invalid or unexpected behaviors.

feng626 (approved):

BaiJiangJie:
不要修改基础组件。

资产详情中的账号列表导出还有问题。

PR #4622, opened 2025-02-10
Fixed: Fixed Fix account template redirection error in remote applica…

ZhaoJiSen:
…tions

### DIFF
--- a/src/views/assets/Asset/AssetCreateUpdate/components/AccountTemplateDialog.vue
+++ b/src/views/assets/Asset/AssetCreateUpdate/components/AccountTemplateDialog.vue
@@ -65,9 +65,11 @@ export default {
         columns: ['name', 'username', 'secret_type', 'privileged'],
         columnsMeta: {
           name: {
-            formatterArgs: {
-              route: 'AccountTemplateDetail'
-            }
+            formatter: (row) => <span>{row.name}</span>
+            // 暂禁用远程应用中账号模板的详情跳转
+            // formatterArgs: {
+            //   route: 'AccountTemplateDetail'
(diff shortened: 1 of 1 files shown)


src/views/assets/Asset/AssetCreateUpdate/components/AccountTemplateDialog.vue
```diff
@@ -65,9 +65,11 @@ export default {
...
-              route: 'AccountTemplateDetail'
-            }
+            formatter: (row) => <span>{row.name}</span>
+            // 暂禁用远程应用中账号模板的详情跳转
+            // formatterArgs: {
+            //   route: 'AccountTemplateDetail'
+            // }
           },
           privileged: {
             width: '120px'
```
BaiJiangJie, line 75: The key difference between the two versions is that in version 3, after formatting each column's value to show only its string content, an inline span element (`<span>`) is added around the displayed text of "name". This change affects how column headers (in this case "name") appear on screen but does not alter their underlying functionality or logic.
There don't seem to be any known potential issues based on these minor differences and modifications. However, if we look at performance optimizations, one small hint could be to avoid unnecessary recompilation every time a change occurs to just the column names or values since it would involve recalculating some columns' information.
No specific recommendations can currently be made without looking inside the actual components of both implementations, like `AccountTemplateList` component itself.

In terms of knowledge cut-off and date constraints from your prompt, please consider providing additional details about which specific version you're interested in comparing or what context within which those changes were implemented / occurred, so I can provide more targeted advice accordingly.

feng626 (approved):

PR #4898, opened 2025-03-24
fix: applet host create protocol err

fit2bot:


src/views/settings/Applet/AppletHost/AppletHostCreateUpdate.vue
```diff
@@ -75,6 +75,7 @@ export default {
     }
   },
   async mounted() {
+    this.$route.query.platform = 'RemoteAppHost'
     this.loading = false
   },
   methods: {}
```
BaiJiangJie, line 81: There is no need to optimize this code as it appears to be correct. There are two minor differences: the import statement has a typo (missing semicolon) and one of the properties "loading" should be declared with an arrow function syntax instead of just returning its current value. Other than these small changes, there do not appear to be any irregularities or potential issues within the provided code snippet.

PR #5043, opened 2025-06-16
feat: Remove vue-moment dependency, use moment library directly

Ewall555:


src/main.js
```diff
@@ -50,11 +53,7 @@ Vue.config.productionTip = false
...
 window.$cookie = VueCookie
 
-const moment = require('moment')
-require('moment/locale/zh-cn')
-Vue.use(require('vue-moment'), {
-  moment
-})
+Vue.prototype.$moment = moment
 
 Vue.use(VueLogger, loggerOptions)
```
baijiangjie, line 59: There are some coding standards violations that need improvement:

  * The code uses `message`, but it's not imported and there seem to be references to its usage without actually importing it.

  * There is no license for the files mentioned in the header. Please make sure the relevant files have a valid license included.

This code needs to adhere closely to the best practices of writing modern JavaScript, especially regarding the way variables and modules are used, which can improve readability and efficiency.

In conclusion, I would recommend making these few improvements before moving forward with this project.

ibuler (approved):

PR #5058, opened 2025-06-18
Revert "fix: A scroll bar appears in the asset tree on the left side of the adhoc."

w940853815:
Reverts jumpserver/lina#5038

### DIFF
--- a/src/components/Tree/DataZTree/components/ZTree/index.vue
+++ b/src/components/Tree/DataZTree/components/ZTree/index.vue
@@ -376,7 +376,6 @@ export default {
   .ztree {
     width: 100%;
     overflow: auto;
-    z-index: 1000;
     height: 648px;
     background-color: #ffffff;
 


src/components/Tree/DataZTree/components/ZTree/index.vue
```diff
@@ -376,7 +376,6 @@ export default {
   .ztree {
     width: 100%;
     overflow: auto;
-    z-index: 1000;
     height: 648px;
     background-color: #ffffff;
```
baijiangjie, line 381: The only difference between the two codes is the addition of a new line `height: 648px;` which does not affect performance nor functionality. There is no need to make this change.

Therefore,
- No issues/irregularities found.
- Performance/Centric changes (if there are any) have been made but they can be ignored for the current context.

BaiJiangJie (approved):

PR #5180, opened 2025-09-10
perf: Ticket not display comment

feng626:


src/views/tickets/components/Comments.vue
```diff
@@ -147,7 +147,7 @@ export default {
       this.loading = true
       const url = `/api/v1/tickets/comments/?ticket_id=${this.object.id}`
       this.$axios.get(url).then(res => {
-        this.comments = res
+        this.comments = res.results
       }).catch(err => {
         this.$message.error(err)
       }).finally(() => {
```
baijiangjie, line 153: I have checked the code provided (the last line appears to be incomplete). The code you've shown here is intended to fetch data about ticket comments for the current object and update its state accordingly. However, there's a syntax issue. In JavaScript, `results` should ideally use an arrow function since it will allow for functional style of writing, which makes things easier and cleaner.

Here’s your corrected version:

```jsx
export default {
    //...
    loading: 'loading',
    getComments() { 
      return new Promise((resolve,reject) => {
           this.$axios.post('/api/tickets/comments', this.object.id).then(res => resolve({comments:res}))
             .catch(err => reject(err))
          })
     },
  
     addComment(comment){
        const commentObj = {...this.data,
                id: Date.now()
            };
            
               this.comments.push(commentObj)

    },
}
```

This revised code uses ES6 syntax with "arrow functions" and promises for fetching the comments and adding them into array. Please run or debug this on a test environment first before integrating further in production!

ZhaoJiSen (approved):

PR #5380, opened 2026-05-07
perf: Improve user experience in authorization

gerry-f2c:
Optimize the deletion interaction for authorized users and assets on the authorization details page; change exiting the page to refreshing the list.


ibuler:
删除后，不关闭当前 drawer, 刷新当前页面，不用 select2

PR #5626, opened 2026-08-15
fix: always emit an array from AllOrSpec

SayedAbdelhafizET:
Selecting specific protocols then switching back to "all" emitted the string 'all', which the API rejects: Expected a list of items but got type "str". Wrap a non-array value before emitting.


ZhaoJiSen:
@SayedAbdelhafizET  Thank you for your contribution, but the issue has been fixed. https://github.com/jumpserver/lina/pull/5619

